PR #2, opened 2019-07-31
Please review BFS-2.

Shazmaan:


Shazmaan:
You could avoid using visited set but I used it to be clear of the concept we discussed yesterday.

PR #4, opened 2019-08-01
Completed BFS -2 - Problem 2 and Problem 3 - Please Review

bharatreddypothireddy:


Problem2.java
```diff
@@ -0,0 +1,81 @@
...
+                }
+                if(j!=0 && grid[i][j-1]==1) {
+                    grid[i][j-1] = 2;
+                    queue.add(i*column+(j-1));
+                    onecount--;
+                }
+                size--;
+            }
+            count++;
+            if(onecount==0) break;
```
super30admin: Due to this line you need not decreease timer by 1.
It takes care of one extra iteration of while loop.

PR #10, opened 2019-08-26
Finished Problem1 & 3 : Please review

muditabysani:


Problem2.py
```diff
@@ -0,0 +1,86 @@
+from collections import deque
+class Solution:
+	def orangesRotting1(self, grid):
+		# Used an approach similar to the game of life problem
+		# But if its not possible to convert everything to stale don't know the stopping condition
```
sehaj-grover: the loop ends when we have iterated through all the stale oranges we have because after that no new oranges can be affected.

Problem2.py
```diff
@@ -0,0 +1,86 @@
...
+			res += 1
+		if count == (rows * columns):
+			return res
+		else:
+			return -1
+
+	def orangesRotting2(self, grid):
+		# Time Complexity : O(mn) where m is the number of rows in the grid and n is the number of columns in the grid
+		# Space Complexity : O(mn) because we are storing every element in the queue
+		# Doubt is at a point we are not storing all the elements in the queue. But I'm not sure.
```
sehaj-grover: We are to just look for the stale oranges since only they are affected the question. So we only store them in a queue

Problem2.py
muditabysani: But in my code I'm looping through all the elements in the grid. So I'm not understanding this iteration over all the stale oranges

PR #22, opened 2019-08-29
SOlutions Added - Please Review

Arpan61:


Problem2.py
```diff
@@ -0,0 +1,35 @@
...
+                    queue.append((r, c, 0))
+
+        while queue:
+            r, c, depth = queue.pop(0)
+            for vr, vc in validindexer(r, c):
+                if grid[vr][vc] == 1:
+                    grid[vr][vc] = 2
+                    queue.append((vr, vc, depth + 1))
+
+        for r in range(rows):
```
sehaj-grover: instead of checking all values again try using a count of fresh oranges left and compare it to 0

Arpan61: ok

PR #25, opened 2019-09-18
Problem 1 solution

siddarth306:


problem2.py
```diff
@@ -0,0 +1,72 @@
...
+            for j in range(len(cost_grid[i])):
+                if cost_grid is not None and cost_grid[i][j] == -1:
+                     return -1
+                else:
+                     if cost_grid[i][j] is not None and max_val < cost_grid[i][j]:
+                        max_val = cost_grid[i][j]
+                     
+                     
+        return max_val
+
```
siddarth306: I forgot to add the time complexity and space complexity. 

Time complexity: O(n^2)
Space complexity: O(n^2)

Algorithm:
Initialize a cost matrix
For every rotten orange, do a BFS where the neighbors will be the top, bottom, left and right oranges. While traversing, update each node with its hop count in the cost matrix. While updating, check if the orange has become rotten already rotten. If yes, ignore that neighbor. 
Do this traversing for each rotten orange. If there is already an orange having a higher hop, replace it with the current rotten orange's lower hop.  
Find the max cost in the cost matrix.

siddarth306:
Yes. Thank you for the feedback. I will work on it.

PR #27, opened 2019-09-18
Add cousins, traversals writeup and code

pradhan1234:
cc @kaurramanpreet
this also contains the write up for traversal bfs and dfs for trees and graphs with its algorithms and codes.


pradhan1234:
added rotten oranges @kaurramanpreet

pradhan1234:
added bipartite graph, which was left to add

PR #41, opened 2019-11-06
Please Review- CousinsInBinaryTree

PrathushaKoouri:
BFS solution did not pass the leetcode - got an internal error message when 'y' is not given.

### DIFF
--- a/59_CousinsInBinaryTree(DFS).py
+++ b/59_CousinsInBinaryTree(DFS).py
@@ -1,0 +1,45 @@
+# Accepted on leetcode(993)
+# time - O(N), space - O(1)
+# We use 4 global variables here xparent and yparent which should not be equal and xdepth and ydepth which should be equal.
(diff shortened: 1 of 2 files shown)


sehaj-grover:
See if this helps
`from Collections import deque
class Solution:
    def isCousins(self, root, x, y):
	    queue = deque()
	    queue.append(root)
	    while len(queue) > 0:
	    	q_size = len(queue)
	    	x_found = False
	    	y_found = False
	    	for i in range(q_size):
	    	    node = queue.popleft()
	    	    if node.left != None and node.right != None:
	    		    if (node.left.val == x and node.right.val == y) or (node.right.val == x and node.left.val == y):
	    			    return False
			    if node.val == x:
	    			x_found = True
	    		elif node.val == y:
	    			y_found = True
	    		if node.left != None:
	    			queue.append(node.left)
				if node.right != None:
	    			queue.append(node.right)
	    	if x_found and y_found:
	    		return True
	    return False
`
else ping me again and we can resolve the issue

sehaj-grover:
Great

PR #48, opened 2019-11-07
Employee importance O(n) solution

sudhasr:


employeeImportance.py
```diff
@@ -0,0 +1,34 @@
...
+            val = importance[id]
+            for i in subordinates[id]:
+                val = val + helper(i) #recursive call for each subordinates 
+            return val   
+        #Creating two ditionaries for storing corresponding employee's imp and their direct subordinates
+        importance = {} 
+        subordinates = {}
+        for i in employees:
+            importance[i.id] = i.importance
+            subordinates[i.id] = i.subordinates
```
sehaj-grover: This can be done in a single dictionary right?

rottenOranges-TLE.py
```diff
@@ -0,0 +1,36 @@
...
+        n = len(grid[0])
+        for i in range(m):
+            for j in range(n):
+                if grid[i][j] == 2:
+                    q.append((i,j))
+                if grid[i][j] == 1:
+                    fresh += 1
+        dirs = ((0,1),(1,0),(-1,0),(0,-1))
+        count = 0
+        while q != None:
```
sehaj-grover: while len(q)>0

sudhasr: Yes. Just understood that while q != 0 leads to an infinite loop. It works when I change to len(q) > 0

employeeImportance.py
sudhasr: Yes. Implemented using single dictionary

sudhasr:
Corrected oranges problem

PR #56, opened 2019-12-25
Adding Problem 3

dasgprat:


Problem1.java
```diff
@@ -0,0 +1,87 @@
...
+ *     TreeNode left;
+ *     TreeNode right;
+ *     TreeNode(int x) { val = x; }
+ * }
+ */
+
+/**
+LeetCode Submitted : YES
+Space Complexity : O(1)
+Time Complexity : O(N)
```
neha79091: are you neglecting the queue used for BFS?

dasgprat: No, this complexity is incorrect. 
Let me fix it.

neha79091: ok

PR #57, opened 2019-12-27
Completed BFS 2

RishikaMachina:


Problem_2.py
```diff
@@ -0,0 +1,41 @@
+# Runs on Leetcode
+
+# BFS
+  # Runtime - O(n)
+  # Memory - O(1)
+
```
neha79091: space complexity will be O(n) as you are using queue

neha79091: good

PR #139, opened 2020-05-26
Create Sample.java. Please review BFS-2

Abinashi7:


iivaishnavii:
Nice to see you have completed all the three problems,You have got the TC, SC, solution right.

PR #208, opened 2020-08-10
BFS2

SantoshReddyKallala:


Cousins.java
```diff
@@ -0,0 +1,38 @@
...
+            yParent = parent;
+            y_depth = level;
+        }
+        _dfs(node.left,level+1,node,x,y);
+        _dfs(node.right,level+1,node,x,y);
+
+
+    }
+    //TIme complexity : o(n)
+    //space complexity : logn -> maximum depth of tree
```
samarthjj, line 36: SC should be O(h), where h is the height of the tree.
log(n) depth is only the case for a perfect/full tree.

EmployeeImportance.java
```diff
@@ -0,0 +1,26 @@
...
+    private void _dfs(int id){
+        Employee temp = map.get(id);
+        result+=temp.importance;
+
+        for(int i=0;i<temp.subordinates.size();i++){
+            _dfs(temp.subordinates.get(i));
+        }
+    }
+    //Time complexity : o(n) -> n is number of employees
+    // space complexity : max depth of tree
```
samarthjj, line 25: O(n), space for the HashMap.

RottenOranges.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        if(freshcount==0){
+            return result-1;
+        }
+
+        return -1;
+    }
+    //Time complexity: o(mXn)
+    //space complexity : o(n)
```
samarthjj, line 48: SC should be O(m*n), what if all cells are added to queue to begin with.

samarthjj:
Good work Santosh, Please review the inline comments. 
Please also try **Employee Importance** Problem with the BFS solution as well.

samarthjj:
Good work Santosh, Please review the inline comments. 
Please also try **Employee Importance** Problem with the BFS solution as well.

PR #259, opened 2020-10-14
bfs 2 Create employee.py

jeevy222:


ramya-sruthi-d (commented):
Good. Explain your approach in comments.

ramya-sruthi-d:
Good. Explain your approach in comments.

PR #545, opened 2021-08-24
Completed all

psambalkar:


Rooppesh:
Good Job on both the problems!

Rooppesh:
Good Job on both the problems!

PR #556, opened 2021-09-14
isCousinBST done: please review

SandhyaGitH:


Pinak08 (commented):
No problem if you took time.
Practice more and logic will be clicked in less time to you.
SC for is cousin will be O(height) (Dfs uses stack internally- you have to consider that as your approach has dfs).

Pinak08:
No problem if you took time.
Practice more and logic will be clicked in less time to you.
SC for is cousin will be O(height) (Dfs uses stack internally- you have to consider that as your approach has dfs).

PR #655, opened 2022-01-21
p1 only

LiberoPat:


brunda4 (commented):
Good work with right side view. Please complete the other problem as well.

brunda4:
Good work with right side view. Please complete the other problem as well.

PR #882, opened 2022-09-14
Completed All 2 problems

Siddhant111:


daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

PR #1025, opened 2023-05-14
completed

vikastadepu:
Check once the space complexities

### DIFF
--- a/Cousins.cpp
+++ b/Cousins.cpp
@@ -1,0 +1,105 @@
+// TC: O(N), SC:O(2^H):
+
+// BFS solution
+class Solution {
+public:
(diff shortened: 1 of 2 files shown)


Cousins.cpp
```diff
@@ -0,0 +1,105 @@
+// TC: O(N), SC:O(2^H):
```
osandeepreddy, line 1: SC is O(N)

Right_view.cpp
```diff
@@ -0,0 +1,68 @@
+
+// TC:O(N), 
+// SC:O(2^H)
```
osandeepreddy, line 3: SC is O(h)

osandeepreddy:
1) SC is O(n)
2) SC is O(h)

PR #1108, opened 2023-09-29
please review BFS-2

anaghayn:


Nandha2Tech (commented):
Great Work! In right side view of a binary tree problem, since queue is used SC would be O(n)

Nandha2Tech:
Great Work! In right side view of a binary tree problem, since queue is used SC would be O(n)

PR #1210, opened 2024-03-09
Completed BFS-2

pkapse173:


Nandha2Tech (commented):
Great Work! At max the size of the queue would be equal to the number of leaf nodes. So SC for both the problems would be O(n/2) -> O(n).

Nandha2Tech:
Great Work! At max the size of the queue would be equal to the number of leaf nodes. So SC for both the problems would be O(n/2) -> O(n).

PR #1277, opened 2024-07-29
completed BFS-2

dhikki-study:
#######199. Binary Tree Right Side View#####################################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem

// Your code here along with comments explaining your approach in three sentences only
I have sued BFS approach but first I will select right child and put it in queue, so for each level the first child is the right most child which will give right view, now to append it in list I made sure only 1st element is inserted in list and rest are ignored

        
        

######993. Cousins in Binary Tree#############################################
// Time Complexity : n
// Space Complexity : O(n)
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : NA

// Your code here along with comments explaining your approach in three sentences only
Here also we have used BFS where before inserting we are checking if they are sibling if that is the case than return False. Next we check if they are presenting in same level and do not have same parents than true or else return false as they will be on different level.

### DIFF
--- a/Sample.py
+++ b/Sample.py
@@ -1,0 +1,99 @@
+#######199. Binary Tree Right Side View####################################################################################################################
+// Time Complexity : n
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : no problem
+
+// Your code here along with comments explaining your approach in three sentences only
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Great Work!

Nandha2Tech:
Great Work!

PR #1330, opened 2024-11-24
COmplete BFS-2

Thaslim:


Erlybird:
In the right side view question, if you are given an empty tree, you must return an empty list, while you are returning null. Please check the base cases.

Erlybird:
In the right side view question, if you are given an empty tree, you must return an empty list, while you are returning null. Please check the base cases.

PR #1363, opened 2025-01-29
BFS-2

SandeepKanumilli:


Nandha2Tech (commented):
Great Work! Space complexity for cousins in binary tree using bfs would be O(n) as in the worst case all the leaf nodes will be there in the queue (n/2 leaf nodes -> O(n))

Nandha2Tech:
Great Work! Space complexity for cousins in binary tree using bfs would be O(n) as in the worst case all the leaf nodes will be there in the queue (n/2 leaf nodes -> O(n))

PR #1505, opened 2025-07-19
solved

sethu95:


super30admin:
Strengths:
- The student demonstrates a good understanding of tree traversal techniques (both DFS and BFS).
- The solutions are efficient and correctly solve the problems.

Areas for Improvement:
- For problem1.java, while the DFS approach is correct, it might be worth noting that BFS is more intuitive for level-order problems like the right side view. However, the DFS approach is equally valid and efficient.
- In problem2.java, the student could add a check to early terminate the recursion once both x and y are found, which could save some unnecessary traversals in large trees.

PR #1516, opened 2025-08-16
problem 1and 2 added

Tejas274:


super30admin:
The student has provided solutions for two problems: Binary Tree Right Side View and Cousins in Binary Tree. Here's the detailed feedback:

1. Correctness:
- For the Binary Tree Right Side View problem, both BFS and DFS approaches are correctly implemented. The BFS approach captures the rightmost node at each level, and the DFS approach updates the result array with the rightmost node encountered at each depth.
- For the Cousins in Binary Tree problem, both DFS and BFS approaches correctly check if nodes x and y are cousins (same depth but different parents). The BFS approach efficiently checks for same-parent conditions during traversal.

2. Time Complexity:
- Both problems correctly state O(n) time complexity where n is the number of nodes, which is accurate for both BFS and DFS traversals.

3. Space Complexity:
- The space complexities are correctly identified as O(n) for BFS (queue size) and O(h) for DFS (recursion stack) where h is the tree height.

4. Code Quality:
- The use of helper methods in DFS approaches improves modularity.
- Some minor improvements could be made:
  - Consistent spacing around operators and after commas
  - Remove duplicate imports (Queue is imported twice in problem1)
  - The second solution in problem1 could be better commented to explain the DFS approach
  - The BFS solution in problem2 could be simplified by tracking parent and depth information in the queue

5. Efficiency:
- The solutions are efficient and use standard traversal approaches.
- For problem2, the BFS solution could potentially exit early once both nodes are found at different depths, but the current implementation is still correct.

PR #1517, opened 2025-08-20
Finished BFS2

sowbaranika1302:


super30admin:
Strengths:
- Good use of DFS and BFS approaches for tree traversal.
- Clear and concise code structure.
- Appropriate use of data structures like deque for BFS.

Areas for Improvement:
- In the BFS solution for rightSideView, ensure you capture the rightmost node by checking the last node in each level (i == size-1) instead of the first node.
- For the isCousins problem, check for siblings more thoroughly by comparing parent nodes, not just when both left and right children exist.
- Improve the return logic in isCousins to handle cases where only one node is found more accurately.
- Avoid using self.res unnecessarily in the BFS solution for rightSideView; a local variable would suffice.

PR #1519, opened 2025-08-20
Done BFS-2

akhilsai25:


super30admin:
- For `BinaryTreeRightView.java`: The current implementation captures the leftmost view of the tree. To fix this, you should capture the last element of each level (when `i == size-1`) instead of the first element (`i == 0`). Also, you should process the left child before the right child to ensure the rightmost node is captured last in each level.
- For `CousinsInBinaryTree.java`: The solution is well-implemented and correctly uses DFS to track the parent and height of the nodes. The logic is clear and efficient.
- General: Your code is clean and well-commented, which is excellent. Just ensure the logic matches the problem requirements before submission.

PR #1521, opened 2025-08-21
added implementations

nsh2207:


super30admin:
Strengths:
- Both solutions correctly implement BFS to solve the problems.
- The isCousins solution effectively uses a hashset to track the presence of x and y.

Areas for Improvement:
- Avoid using sentinel nodes like -1000; instead, track the level size directly as shown in the reference solution.
- Improve variable naming for clarity (e.g., 'sep' could be named 'levelMarker').
- Add more detailed comments to explain the logic, especially for the isCousins solution.
- Consider separating the TreeNode definition into a header file for better code organization.

PR #1525, opened 2025-08-29
BFS-2 Done

Sarthak15997:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, with optimal time and space complexity.
  - The use of DFS is appropriate for both problems, especially for "Cousins in Binary Tree."

- Areas for Improvement:
  - For "Binary Tree Right Side View," consider traversing right first and then left. This would allow you to append the first node encountered at each level (the rightmost node) directly to the result list without overwriting. This would make the logic clearer and slightly more efficient.
  - While the space complexity for "Binary Tree Right Side View" is better than the reference solution, the BFS approach is more intuitive for this problem. Consider explaining why you chose DFS over BFS in your comments.

PR #1532, opened 2025-09-18
Completed both the problems

restacksyj:


super30admin:
- Strengths:
  - Both solutions correctly implement BFS and handle edge cases well.
- Areas for Improvement:
  - Avoid using `<bits/stdc++.h>` in favor of specific headers like `<queue>` and `<vector>` for better practice.
  - In Problem2.cpp, consider breaking the loop early once both nodes are found or if their parents are the same to improve efficiency.
  - The logic in Problem2.cpp could be slightly more concise by combining some conditions.

PR #1539, opened 2025-09-19
Add solutions for binary tree right side view and cousins

ravithakur2k:
Added BinaryTreeRightSide.py implementing BFS to get the right side view of a binary tree, and CousinsBinaryTree.py using DFS to determine if two nodes are cousins by comparing their depths and parents.


super30admin:
- Strengths:
  - The student has a good understanding of BFS and DFS and has applied them correctly to solve the problems.
  - The student has correctly analyzed and stated the time and space complexities.
- Areas for Improvement:
  - For the right side view problem, the student could consider mentioning that the BFS approach ensures we visit nodes level by level, which is why we can capture the rightmost node at each level.
  - For the cousins problem, the student could add a brief explanation of why the DFS approach is suitable (e.g., it allows us to track both parent and depth simultaneously).
  - The student could also consider adding edge case checks (e.g., if x or y is not found in the tree) to make the solution more robust, although the current implementation handles these cases implicitly.

PR #1541, opened 2025-09-20
Implemented BFS-2

johar-inderpreet:
Implemented BFS-2


super30admin:
- The main issue with your solution is the incorrect logic for determining the right side view. You should be adding the last element of each level (i == size-1) to the output list, not the first element (i == 0). This is because the right side view requires the rightmost node at each level, which is the last node processed in a left-to-right level order traversal.
- The rest of your code is well-written and follows good practices. The comments and variable names are clear, and the structure is easy to follow.
- For future problems, ensure you thoroughly test your logic with different tree structures to verify correctness before finalizing your solution.

PR #1546, opened 2025-10-02
BFS-2

v-hemish:


super30admin:
- problem1.py: Well done! Your solution correctly implements the right side view using DFS. The logic is clear and efficient. One minor suggestion is to rename `mp` to something more descriptive like `right_view` to enhance readability.
- problem2.py: Your approach is on the right track, but there's a logical flaw in checking for cousins. You should only check if both nodes are found at the end of processing each level, not within the loop. Also, consider using more descriptive variable names to improve code readability. Here's a corrected version of the BFS loop:
  ```python
  while q:
      size = len(q)
      found_x = False
      found_y = False
      px = None
      py = None
      for _ in range(size):
          node = q.popleft()
          pnode = pq.popleft()
          if node.val == x:
              found_x = True
              px = pnode
          if node.val == y:
              found_y = True
              py = pnode
          if node.left:
              q.append(node.left)
              pq.append(node)
          if node.right:
              q.append(node.right)
              pq.append(node)
      if found_x and found_y:
          return px != py
      if found_x or found_y:
          return False
  return False
  ```

PR #1547, opened 2025-10-08
Completed

shantanujoshi25:


super30admin:
- **Problem_1.py**: Well done! Your solution correctly implements the BFS approach to find the right side view. The code is clean and efficient.
- **Problem_2.py**: Your approach to the cousin nodes problem needs improvement. The logic for determining cousins is flawed. Instead of tracking indices in a list, you should track the parent and depth of each node. Consider using a BFS approach where you keep track of the parent and depth for each node, then compare these values for nodes x and y. This will give you a more accurate and efficient solution.

PR #1550, opened 2025-10-25
BFS-2 Done

JPolishetty:


super30admin:
- Strengths:
  - Both solutions are correct and efficient, matching the reference solutions in time and space complexity.
  - The use of BFS for Problem1 and DFS for Problem2 is appropriate and correctly implemented.

- Areas for Improvement:
  - In Problem2, consider renaming variables like `lvl` to `level` for better readability.
  - While the comments are helpful, they could be slightly more concise in Problem1 (e.g., "AS the right side view means last element in each level" could be simplified to "Right side view is the last node at each level").

PR #1551, opened 2025-10-27
Completed BFS-2

nikhilsaimarri:


super30admin:
Strengths:
- The code is well-organized and readable.
- The use of recursion and helper methods is a good practice.
- The comments provide clear explanations of the intended logic.

Areas for Improvement:
- The logic for capturing the rightmost node is flawed. The current approach updates the result list with the most recently seen node, which does not necessarily correspond to the rightmost node at each level. Instead, you should process nodes level by level and capture the last node in each level, as done in the reference solution.
- Consider using an iterative approach with a queue for level-order traversal, which is more straightforward for this problem.
- Ensure that the solution's logic aligns with the problem requirements before focusing on optimizations.

PR #1553, opened 2025-10-30
Cousins in Binary Tree & Right Side view of tree

ashishbadhe97:


super30admin:
The student's solution is excellent. However, there are a few minor points for improvement:
- In the DFS approach, the comment "since we start our level from 0, if result size is equal to level, we dont have elements sotred for that level" could be clearer. Consider rephrasing for better readability.
- In the BFS approach, the comment "size to remove only those elements which are on current level" could be more precise. For example, "size represents the number of nodes at the current level, ensuring we process only those nodes."

Overall, the student has demonstrated a strong understanding of the problem and provided efficient, correct solutions.

PR #1557, opened 2025-11-10
bfs-2

ankitakulkarnigit:


super30admin:
Strengths:
- Both solutions correctly implement BFS and solve the problems as intended.
- The logic for identifying cousins is sound and correctly handles edge cases.

Areas for Improvement:
- For Problem 1, replace the list `q` with `deque` from the `collections` module to improve efficiency of `popleft` operations.
- In Problem 2, consider tracking parent nodes explicitly to make the sibling check more straightforward.
- The comments, while helpful, could be more concise to avoid clutter.
- The variable names like `last_ele` could be more descriptive (e.g., `rightmost_node`).

PR #1568, opened 2025-12-31
Solved BFS-2 problems

aahiawasthee:


super30admin:
- For "binary-right-side.java":
  - The approach needs to be corrected to capture the rightmost node at each level. Instead of adding the right child first and checking i==0, you should add the left child first and check i==size-1 to get the rightmost node.
  - Correct the space complexity comment to O(n) as the queue can hold up to n nodes in the worst case.

- For "cousins-in-binary.java":
  - The logic is almost correct, but you should ensure both x and y are found at the same depth before returning true. Currently, if one is found and the other is not, it returns false immediately, which might not always be correct.
  - Consider adding a check to ensure both x and y are found in the same level before concluding they are cousins.

- General:
  - Always double-check your logic against edge cases to ensure correctness.
  - Pay attention to space complexity comments, especially when using data structures like queues.

PR #1572, opened 2026-01-21
Completed BST 2

thondaha:
Right side view of a tree
Cousins in Binary tree


super30admin:
The student's solution is excellent overall. Here are some minor suggestions for improvement:
- The DFS approach could include a comment explaining why the right child is processed after the left child, ensuring the rightmost node is always the last one processed at each level.
- The test cases could be expanded to include edge cases such as an empty tree or a tree with only one node to ensure comprehensive testing.
- The helper method in the DFS approach could be made private to enforce encapsulation, although this is a minor point.

PR #1575, opened 2026-01-24
BFS2-completed

sujaygijre:


super30admin:
Your solution is well-implemented and correct. You have correctly used a queue to perform level order traversal and captured the last node of each level. This is the standard approach for this problem.

One minor suggestion: instead of using `while (size--)` and checking `if (size == 0)`, you could use a for loop that iterates exactly `size` times and check when the index is `size-1`. This might make the intention slightly clearer, but your current approach is also acceptable.

Note that you included an unrelated file (cousingInBinaryTree.cpp) in your submission. Make sure to only submit the code for the problem you are solving.

Overall, great job!

PR #1577, opened 2026-01-25
BFS-2 complete

Nikesh46:


super30admin:
Your solution for the Right Side View problem is correct and efficient. Well done! However, note that you included code for a different problem (Leetcode993) in the same file. In the future, make sure to submit only the code for the problem you are solving.

To improve code clarity:
1. Consider renaming "q" to "queue" for better readability.
2. The inner while loop can be replaced with a for loop that runs from 0 to size-1, which might make the intention clearer (as in the reference solution).
3. The comment at the top says "DFS Solution", but your code uses BFS. Please correct the comment to avoid confusion.

Here's a slight refactor of your code for clarity (without changing the logic):
```
class Solution {
    public List<Integer> rightSideView(TreeNode root) {
        List<Integer> res = new ArrayList<>();
        if (root == null) return res;
        Queue<TreeNode> queue = new LinkedList<>();
        queue.add(root);
        while (!queue.isEmpty()) {
            int size = queue.size();
            for (int i = 0; i < size; i++) {
                TreeNode node = queue.poll();
                if (i == size - 1) {
                    res.add(node.val);
                }
                if (node.left != null) {
                    queue.add(node.left);
                }
                if (node.right != null) {
                    queue.add(node.right);
                }
            }
        }
        return res;
    }
}
```
This uses a for loop which is more common for level-order traversal.

Nikesh46:
Quick Question - Leetcode993 contains isCousins problem which was one of the questions in BFS-2. Please let me know if I am missing something here?

PR #1580, opened 2026-01-26
2 problems completed

BharathVuppala96:


super30admin:
Your solution for the "Right Side View" problem is correct and follows the standard BFS approach. Well done! However, there are a few areas for improvement:

1. **Efficiency of Queue Operations**: You are using a list for the queue and performing `pop(0)` which is an O(n) operation. This makes your solution less efficient for large trees. Instead, you should use `collections.deque` which allows O(1) popleft operations. Here's how you can modify your code:
   ```python
   from collections import deque
   # ...
   q = deque([root])
   while q:
        s = len(q)
        for i in range(s):
            curr = q.popleft()
            # ...
   ```

2. **Variable Naming**: While short variable names are convenient, using more descriptive names can make your code easier to understand. For example, `q` could be renamed to `queue`, `s` to `level_size`, and `curr` to `current_node`.

3. **Edge Handling**: You correctly handle the case when the root is None. That's good!

4. **Unnecessary Code**: The solution is concise and to the point. There is no unnecessary code.

For the "Cousins in Binary Tree" solution, note that it was not required for this problem. However, if you are submitting multiple solutions, ensure they are in separate files and named appropriately.

Keep up the good work! With a small change to use `deque`, your solution will be efficient and optimal.

PR #1581, opened 2026-01-27
Done with BFS-2

Mukul585:


super30admin:
Your solutions for the "Right Side View" problem are correct and efficient. Well done! You have implemented both BFS and DFS approaches correctly.

Strengths:
- You handled the base case (null root) appropriately in the BFS solution.
- Both solutions have optimal time and space complexity.
- The code is readable with clear variable names.

Areas for improvement:
- Please ensure that you only submit code relevant to the problem. The inclusion of "Cousins in Binary Tree" solutions in the same file is confusing and might be an error. In the future, separate solutions for different problems into distinct files.
- For the DFS solution, consider adding a brief comment explaining that by traversing the right subtree first, we ensure that the first node we encounter at each depth is the rightmost one.
- Although not necessary, you could make the DFS solution slightly more robust by adding a null check for the root in the public method (as you did in the BFS solution) before calling dfs. However, the current code handles it correctly because the dfs method checks for null.

Overall, your understanding of tree traversals is evident. Keep up the good work!

PR #1582, opened 2026-02-03
Adding BFS-2 solutions

subbu4061:


super30admin:
Strengths:
- You provided two correct solutions for the "Right Side View" problem, demonstrating knowledge of both BFS and DFS approaches.
- The code is clean and well-commented, with explanations of time and space complexity.

Areas for Improvement:
- Ensure that you only submit code relevant to the problem. The solution for "Cousins in Binary Tree" should not be included in the same file for the "Right Side View" problem.
- In the BFS solution, the check `if(curr!=null)` is redundant because you only add non-null nodes to the queue. You can remove this check to simplify the code.
- For the DFS solution, consider adding a base case check at the beginning of the helper method to avoid unnecessary recursion.

Overall, your solutions for the "Right Side View" problem are correct and efficient. However, the inclusion of unrelated code might indicate a need for better organization.

PR #1583, opened 2026-02-05
BFS 2 done

Tanmay-Nalawade:


super30admin:
No specific feedback provided. Refer to GitHub Comment.

super30admin:
Strengths:
- You have provided solutions for both problems and included comments about time and space complexity.
- Your BFS solution for the right side view is correct and efficient.
- You have provided alternative solutions (DFS for right side view and two approaches for cousins) showing your ability to think of multiple approaches.

Areas for Improvement:
- For the DFS solution of the right side view, the approach is incorrect. The standard DFS approach for right side view should prioritize the right subtree first. Then, when we reach a new level (i.e., when the level equals the length of the result list), we add the node value. We should not overwrite the value at a level. Instead, we should only add the first node we encounter at each level when traversing from the right. Here is a corrected DFS version:

        def helper(node, level, res):
            if not node:
                return
            if level == len(res):
                res.append(node.val)
            helper(node.right, level+1, res)
            helper(node.left, level+1, res)

   Notice that we traverse right first and then left. This ensures that the first node we see at a level is the rightmost. Then we only add one node per level.

- In your DFS solution, you are overwriting the value at the level, which means that the last node visited in that level (which might not be the rightmost) is stored. For example, if you have a tree with a left subtree that has more depth than the right, the overwriting might store a node from the left subtree at a deeper level.

- For the cousins problem, both solutions are correct. However, the second solution might be less straightforward because it checks for siblings at the same parent during the traversal. It is efficient, but the first solution (storing the parent explicitly) is more clear.

Recommendation:
- Always test your solutions with various test cases, especially edge cases and asymmetric trees.
- For DFS solutions on tree problems, consider the order of traversal (right first vs left first) to achieve the desired result.

PR #1584, opened 2026-02-09
Complete BFS-2

paridhimalviya:

### DIFF
--- a/CousinsInTheTree.swift
+++ b/CousinsInTheTree.swift
@@ -1,0 +1,256 @@
+//
+//  CousinsInTheTree.swift
+//  DSA-Practice
+//
+//  Created by Paridhi Malviya on 1/22/26.
+//
+
+
+/*
+ Using BFS, when we take any element out of the queue, then chekc if it matches any of the element. If yes,
+ check id they are from the same level - using size variable
+ At 1 level, if we find only one node, then return false.
+ If both nodes at a level, check parents if they are equal or not
+ time compelxity - O(n), space compelxity -> O(2n) = O(n)
+ */
+
+
+
+class CousinsInTheTree {
+    
+    final class QueueUsingLL<T> {
+        
+        private final class LinkedListNode<T> {
+            var value: T
+            var next: LinkedListNode<T>?
+            
+            init(value: T) {
+                self.value = value
+            }
+        }
+        
+        private var front: LinkedListNode<T>?
+        private var rear: LinkedListNode<T>?
+        private var count: Int = 0
+        
+        //MARK: - Enqueue
+        func enqueue(_ value: T) {
+            let newNode = LinkedListNode(value: value)
+            if (rear == nil) {
+                front = newNode
+                rear = newNode
+            } else {
+                rear?.next = newNode
+                rear = newNode
+            }
+            count += 1
+        }
+        
+        //MARK: dequeue
+        @discardableResult
+        func dequeue() -> T? {
+            guard let frontNode = front else {
+                return nil
+            }
+            
+            let value = frontNode.value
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- You have implemented multiple approaches (DFS and BFS) to solve the problem, demonstrating a good understanding of different traversal techniques.
- The DFS solution that processes the right child first is elegant and efficient.
- The code is generally well-commented and structured.

Areas for Improvement:
1. **Code Organization**: The `RightViewOfTree` class contains multiple methods for the same problem. It would be better to focus on one solution per problem unless the problem requires multiple approaches. For production code, choose the most efficient or readable solution.
2. **Queue Implementation**: While your custom queue is correct, it is defined in a different file and might not be necessary. Consider using Swift's built-in array with careful indexing or a more efficient queue implementation from a standard library if available. Alternatively, you can use an array and remove elements from the front, but note that this is O(n) per dequeue operation. For BFS, the number of operations is O(n), so using an array would lead to O(n^2) time complexity. Therefore, your linked list queue is better for performance.
3. **Error in DFS Method**: The method `dfs2` in `rightSideView2` calls `dfs` instead of `dfs2` in the recursive calls. This would cause a compilation error or infinite recursion. Correct this if you intend to use that method.
4. **Naming and Clarity**: The method names and variable names are generally clear. However, the second DFS method (`rightSideView2`) and its helper (`dfs2`) are not consistently named. Ensure that helper methods are correctly referenced.
5. **Redundancy**: The `result2` variable is declared as a class property but is only used in `rightSideView2`. It would be better to declare it locally within the method or avoid using a property to store state for DFS to make the code more reusable and thread-safe.

Overall, your solutions are correct and efficient. The main issue is the typo in `dfs2`, but since it is not used in the primary methods, it does not affect the correctness of the `rightSideView` and `rightSideViewBFS` methods.

super30admin:
Your solution demonstrates a good understanding of both DFS and BFS approaches for solving the right side view problem. Here are some strengths and areas for improvement:

Strengths:
- You implemented multiple approaches (DFS and BFS), which shows a deep understanding of the problem.
- The DFS method that prioritizes the right child first is efficient and correct.
- The BFS method is correctly implemented and follows the standard level order traversal.

Areas for Improvement:
1. **Bug in second DFS method**: In `rightSideView2`, you call `dfs` instead of `dfs2`. Also, in `dfs2`, you should use the `result` parameter instead of the class property `result2`. To fix this, change:
   - In `rightSideView2`, call `dfs2(root, level: 0, result: &result2)`.
   - In `dfs2`, replace `result2` with the parameter `result`.
2. **Queue Implementation**: The BFS method relies on a custom queue (`QueueUsingLL`) that is defined in another class. For clarity and to avoid dependencies, you should either define the queue in the same file or use a standard data structure. In Swift, you can use an array to simulate a queue (though it may not be efficient for large deques) or use a more efficient data structure. Alternatively, you can use `Array` with `removeFirst()` (but note that this is O(n) operation) or use a double-ended queue from a library. However, for trees with up to 100 nodes, it is acceptable.
3. **Code Organization**: Consider separating the different solutions into distinct functions or files to avoid confusion. Also, ensure that each function is self-contained and does not rely on class properties unless necessary. For example, the DFS methods use class properties for the result, which is not ideal. Instead, you can use a local variable and pass it by reference (as you did in the first DFS method).
4. **Comments and Readability**: The code is readable, but you should add comments to explain the logic of each approach. For instance, in the DFS methods, explain why you are appending or overwriting values at each level.

Overall, your solution is mostly correct, but the second DFS method has a bug that needs fixing. The BFS method is correct but depends on an external queue implementation.

PR #1588, opened 2026-03-04
Complete BFS-2

PrakarshKamal:


super30admin:
Your solution for the "Right Side View" problem is excellent. It correctly uses BFS to traverse the tree level by level and adds the last node (rightmost) of each level to the result. The code is clear and efficient.

However, note that you included an additional solution for "Cousins in Binary Tree" in the same file. This is not required for the current problem and should be removed to keep the submission focused. Always ensure that your submission only contains the solution for the problem at hand.

For the "Right Side View" solution, you might consider adding a comment to explain why you are adding the last node (i.e., because it's the rightmost node at that level). But the code is self-explanatory.

Keep up the good work!

PR #1591, opened 2026-03-06
Completed BFS-2

sarvanibaru:


super30admin:
Your solution for the "Right Side View" problem is excellent. It correctly implements the level-order traversal and efficiently captures the rightmost node at each level. Your code is clean and well-commented. Keep up the good work!

PR #1597, opened 2026-03-18
Problem1 and Problem2 added

megharaykar:


super30admin:
### Right Side View of a tree (right-side-view.py)
Your solution is correct and efficient. You have successfully implemented a BFS approach to capture the rightmost node at each level. Here are a few points for improvement:
1. Remove the print statement (print(size)) as it is not needed and can slow down the code in large inputs.
2. The condition for appending the value (if i == size-1) is placed after enqueuing the children. While this does not change the correctness, it is more common to check the condition before enqueuing children. However, both ways are acceptable.
3. Consider using a more descriptive variable name than "qroot" (e.g., "node") to improve readability.

Overall, you have a good understanding of the problem and have implemented a solution that meets the required time and space complexity.

VERDICT: PASS

---

### Cousins in Tree (cousins.py)
Your solution is well-structured and correctly implements the BFS approach. Here are some suggestions for improvement:
1. Variable Names: Consider using more descriptive names. For example, instead of `qroot` and `pqroot`, you could use `current_node` and `current_parent` to make the code clearer.
2. Condition Checks: You can simplify the condition checks at the end of the level. The reference solution does:
   - If both x and y are found, return true if parents are different (otherwise false? but note: the reference returns immediately if both found and parents are different, but if they are the same it would not return until the next level? Actually, the reference checks at the end of the level: if both found, then return whether parents are different. Otherwise, if only one found, return false. Your code has an explicit check for same parent which is correct, but note that if both are found and parents are the same, you return false. However, in the reference, it returns false only if one is found? Wait, let me clarify: in the reference, after processing a level, if both are found, it returns true if parents are different. Otherwise, if only one is found, it returns false. But if both are found and parents are the same, it would not return at that level? Actually, no: the reference code has:
        if(x_found && y_found) return x_parent != y_parent;
        if(x_found || y_found) return false;
   So if both are found and parents are the same, it returns false. Your code does the same but with an extra condition. However, your code returns false for same parent and true for different parents. But note: your code has:
        if x_found and y_found and x_parent != y_parent: return True
        elif x_found and y_found and x_parent == y_parent: return False
        elif x_found or y_found: return False
   This is correct and equivalent to the reference. However, you can write it more concisely as:
        if x_found and y_found:
            return x_parent != y_parent
        if x_found or y_found:
            return False
   This avoids the extra elif and is cleaner.

3. Base Case: Your base case (if root is None) is correct, but note that the problem states the tree has at least 2 nodes, so it might not be necessary. However, it's good practice.

Overall, your solution is correct and efficient. With minor improvements in code clarity, it would be excellent.

VERDICT: PASS

PR #1598, opened 2026-05-07
solution BFS 2

shakthinandana:


super30admin:
### Right Side View of a tree (binary-tree-right-side-view.py)
Your solution is well-implemented and correct. Here are some strengths and areas for improvement:

Strengths:
- You have correctly identified the time and space complexity.
- The code is concise and easy to understand.
- The recursive approach is appropriate for this problem.

Areas for improvement:
- Although the code is clear, adding a brief comment explaining the strategy (e.g., "We traverse right first to ensure we capture the rightmost node at each level") would make it even better.
- Consider mentioning that the space complexity due to recursion is O(h) and that in the worst case (skewed tree) it becomes O(n), which is acceptable given the constraints.
- While the solution is efficient, note that for extremely large trees (beyond the constraints), recursion might lead to stack overflow. However, given the constraints (up to 100 nodes), this is not an issue.

Overall, this is a solid solution that meets the problem requirements.

VERDICT: PASS

---

### Cousins in Tree (cousins-in-binary-tree.py)
Your solution has a good structure and uses BFS which is appropriate for this problem. However, there are critical flaws in the logic:

1. **Parent Tracking**: You are only setting the parent when the current node has a child with value x or y. This means that if x or y is the root, it will never be detected because the root has no parent. The root should be handled separately.

2. **Depth Tracking**: Your code does not explicitly track the depth of the nodes. The problem requires that the two nodes have the same depth. Your current approach processes one level at a time, which is good, but you need to ensure that both nodes are found at the same level.

3. **Node Existence**: The problem states that both x and y exist in the tree, so you don't have to worry about missing nodes. However, your code might return false prematurely if only one node is found in a level.

4. **Improvement**: You should modify your approach to record the depth and parent for each node when it is encountered. Alternatively, you can use the reference solution's approach: for each level, check if both nodes are found. If both are found in the same level, then check if they have different parents. If only one is found in a level, return false immediately because they are not at the same depth.

5. **Suggestion**: Instead of only checking the children for x and y, you should check the value of the current node. Also, you should store the parent for each node when you enqueue its children. The reference solution uses two queues: one for the nodes and one for their parents. You can do something similar in Python by storing tuples (node, parent) in the queue.

Here is a corrected version of your code:

```python
from collections import deque

class Solution(object):
    def isCousins(self, root, x, y):
        queue = deque()
        queue.append((root, None))  # (node, parent)
        while queue:
            level_size = len(queue)
            x_parent = None
            y_parent = None
            for _ in range(level_size):
                curr, parent = queue.popleft()
                if curr.val == x:
                    x_parent = parent
                if curr.val == y:
                    y_parent = parent
                if curr.left:
                    queue.append((curr.left, curr))
                if curr.right:
                    queue.append((curr.right, curr))
            if x_parent is not None and y_parent is not None:
                return x_parent != y_parent
            if x_parent is not None or y_parent is not None:
                return False
        return False
```

This way, you are storing the parent for each node when you enqueue it. Then, when you process a level, you check if both x and y are found in that level. If both are found, you check if their parents are different. If only one is found, you return false because they are not at the same depth.

VERDICT: NEEDS_IMPROVEMENT

PR #1603, opened 2026-05-12
Completed BFS-2 Assignment

Hriday-A:


super30admin:
### Right Side View of a tree (tree_right_side_view.java)
**Strengths:**
- Good understanding of the problem, implementing DFS as an alternative to BFS
- Clear comments explaining the logic (why we start from right, why we check size == level)
- Proper handling of edge cases (null root returns empty list)
- Good variable naming and self-documenting code

**Areas for Improvement:**
- Code formatting needs attention - inconsistent indentation (e.g., `// 1.`, `// 2.` comments are not aligned properly)
- Consider adding a base case check at the beginning of the helper method for consistency
- The commented BFS code at the top could be removed for cleaner submission, or kept in a separate file
- Could add a brief explanation of why DFS (right-first) works for this problem

VERDICT: PASS

---

### Cousins in Tree (cousins_in_binary_tree.java)
**Strengths:**
1. Good understanding of BFS approach for level-order traversal
2. Clear variable naming convention
3. Good attempt to optimize space by removing parent queue
4. Well-commented code explaining the approach

**Areas for Improvement:**
1. **Critical Bug**: The second `if(xp!=null && yp!=null) return false;` is unreachable code. The logic should be `if(xp!=null || yp!=null) return false;` to handle the case where only one node is found at the current level.
2. **Logic Error**: When only one of x or y is found at a level (but not both), the function should return false immediately. The current code doesn't handle this correctly.
3. **Unreachable Code**: The second condition can never be true because if both are non-null, the first condition already returns.
4. **Testing**: The first commented-out version is actually correct. Consider testing both approaches.

**Recommendation:**
Fix the condition from `if(xp!=null && yp!=null) return false;` to `if(xp!=null || yp!=null) return false;` to properly handle the case where only one node is found at the current level.

VERDICT: NEEDS_IMPROVEMENT

PR #1606, opened 2026-05-18
Binary Tree Right Side View and Cousins in Binary Tree

anirudhv98:
Completed BFS-2, please review.


super30admin:
### Right Side View of a tree (Problem1.cs)
**Strengths:**
- Correct DFS approach with right-first traversal, which is an elegant solution to this problem
- Clear and concise comments explaining the approach
- Better space complexity than the reference solution in most cases (O(h) vs O(n))
- Clean, readable code with good variable naming

**Areas for Improvement:**
- Consider passing `result` as a parameter to the helper method instead of declaring it as a class-level field for better encapsulation
- Could add null check at the beginning of `RightSideView` method for consistency with the reference solution, though the current implementation handles null implicitly

Overall, this is a solid solution that demonstrates good understanding of tree traversal and the problem requirements.

VERDICT: PASS

---

### Cousins in Tree (Problem2.cs)
**Strengths:**
- Good use of DFS with level tracking
- Clear comments explaining the approach
- Efficient space complexity in balanced trees (O(log n) average vs O(n) for BFS)

**Areas for Improvement:**
- The initial recursive calls miss the root node itself - add a check for root.val == x or root.val == y at the start
- Consider adding early termination when both xParent and yParent are found to avoid unnecessary traversal
- The Helper function signature is awkward - passing both parent and root when parent can be derived is unnecessary
- Could use a cleaner single recursive function that takes just the current node and its parent

**Suggested Fix:**
```csharp
public bool IsCousins(TreeNode root, int x, int y) {
    Helper(root, null, x, y, 0);
    return xLevel == yLevel && xParent != yParent;
}

public void Helper(TreeNode root, TreeNode parent, int x, int y, int level) {
    if(root == null) return;
    
    if(root.val == x) {
        xLevel = level;
        xParent = parent;
    }
    if(root.val == y) {
        yLevel = level;
        yParent = parent;
    }
    
    // Early termination
    if(xParent != null && yParent != null) return;
    
    Helper(root.left, root, x, y, level + 1);
    Helper(root.right, root, x, y, level + 1);
}
```

VERDICT: NEEDS_IMPROVEMENT

